PR #2, opened 2020-03-03
Better quality!

pjlao307:
or is it?


geohot:
Still wavy where the road meets the edge of the road. And the dots should be labelled as lane markings. But a lot better than what we had!

PR #6, opened 2020-03-03
0005

mart9n:


geohot:
Can you check that you only used the 5 color palette? My script is showing there's other colors in the image

PR #8, opened 2020-03-03
Include some better instructions

techied:
To help the less technically-inclined


techied:
"The Idea" section now redundant since comma pencil is included?

PR #17, opened 2020-03-03
Converted 5 imgs to masks

lambaek:
Converted 5 imgs to masks and deleted the 5 imgs from /imgs


geohot:
Oh, don't delete the images! But masks look great!

PR #18, opened 2020-03-03
101 102 103 104 105

lambaek:
imgs converted to masks


geohot:
The wire in the upper right hand corner of 253 should be labeled "my car"

geohot:
Also, be careful that only the 5 colors are used. I see others in some of the images

PR #21, opened 2020-03-03
Updated Masks

JamesCCockram:


geohot:
Thanks for the PR!

0008 has some lane junk on the right side
0011 has a wobbly road
0012 has a speck of road in the middle of nowhere

geohot:
Also, 0015 shouldn't have the left turn arrow as a "lane marking"

JamesCCockram:
I'll quickly update them now

JamesCCockram:
0008, 0011, 0012 and 0015 should be fixed now

geohot:
I think a few of these could still use some refinement, but it's a whole lot better than what's in there. Going to merge :)

It's important to be pixel accurate if possible, each pixel is effectively a classification problem for training the network, and wrong pixels confuse it.

geohot:
Crosswalk shouldn't be labelled as "lane marking" in 15. Masks should be much tighter in 11.

PR #22, opened 2020-03-03
[455-460]

mobula3967:


geohot:
In 0455, the left turn arrow shouldn't be including in "lane markings"

VirtuallyChris:
Thanks for your contribution! I'm seeing some anti-aliasing artifacts in the images. You'll need them to be hard edged.

image

I'd recommend going through these again to clean the edges up. You can make sure this doesn't happen again by following these tips:
- instead of Brush use Pencil
- In selection tools turn off anti-aliasing in the options bar

pjlao307:
For 456 need to mark the entire lane line (black and white).
For 458 the road label is overlapping the undriveable (barrier) on the left side towards the horizon.
The rest look great!

Hubblesphere:
Looks good!

PR #30, opened 2020-03-04
0101, 0102, 0103, 0104, and 0105

jason-h-simplifi:
Done with 0101 and 0102.  I plan to do all of these.


jason-h-simplifi:
Done with 0103

pjlao307:
For 102, looks like you're missing part of the hood on the right side.
For 103, that truck in the far distance is missing the label.

jason-h-simplifi:
Fixed!  Lmk if you've got more feedback.

pjlao307:
These look good!

PR #31, opened 2020-03-04
Update 0076_b8727c7398d117f5_2018-11-10--11-05-38_23_757.png

FastCashHash:
How does this look?


pjlao307:
The left turn only marking should not be labeled as a lane.  Should just be part of the road.  
Also the mask on that tractor looks a bit off.

FastCashHash:
Ok. Thanks for the feedback. I will try to address these issues in the morning. Is there a better way than “eyeballing” it?

pjlao307:
You'd need to zoom in to see the contours of the object and be able to select them.  That's how I trace them using photoshop, extensive use of zoom.

pjlao307:
Looks like you have some color issues with 70,71,76.

pjlao307:
Why does this PR have this file?
imgs/0071_a60668e197b93199_2018-08-23--13-17-24_29_1107.xcf

pjlao307:
70 now has conflicts.  I had already submitted an update for 70.  Can you remove your update for that file?  You can do that by copying that file from comma's master branch (https://github.com/commaai/comma10k/raw/master/masks/0070_f550e313a9262051_2018-06-25--09-55-16_2_784.png) into yours and committing that.

pjlao307:
Also looks like you have bad colors in them that you'll need to fix.

pjlao307:
Is this PR still being worked on?

PR #39, opened 2020-03-04
80, 81, 82 new masks

Hubblesphere:


geohot:
I can see the lane lines in 82, we should label even very faint ones.

PR #52, opened 2020-03-05
0317 add lanes

dorkmo:
first edit. i dont know if this helps anything


VirtuallyChris:
Good start! This image could use cleaner identification of the cars and road as well.

dorkmo:
added some more detail, the road edge on the right side is a bit ambiguous, goes from paved, to rocky, to grass.

PR #50, opened 2020-03-05
Updates Masks for 150 - 153

scottadair:

PR #57, opened 2020-03-05
Revert "Updates Masks for 150 - 153"

VirtuallyChris:
Reverts commaai/comma10k#50 - still bad colors :c


scottadair:
Get out of town.. Which file?

PR #69, opened 2020-03-05
Updated mask for 0237_8443d7dd0ec9bd92_2018-08-31--22-04-01_16_300

brittonx:


VirtuallyChris:
This image has anti-aliasing all over it making it have more than 5 colors. Use the pencil tool and disable anti-aliasing when doing selections and fills so the edges are hard.

pjlao307:
The file still has bad colors that need to be fixed.

pjlao307:
Is this still being worked on?

brittonx:
Hi!
Yes, I should get through this one and the others I claimed in the next day
or two.

pjlao307:
I'm going to close this one so someone else can grab it.

PR #76, opened 2020-03-05
Pencil basic windows compatibility

nelsonjchen:
The bundled PEP8 reformatting is probably reaching a bit far and is pretty porky. 

I haven't tested if commits or anything else work. I do know that the images now load and editing can at least be done.

### DIFF
--- a/pencil/server.py
+++ b/pencil/server.py
@@ -4,73 +4,84 @@ import glob, json, subprocess, os
 
 app = Flask(__name__, static_url_path='')
 IMAGES = sorted(glob.glob('../imgs/*.png'))
-config = json.load(open('config.json','r'))
+config = json.load(open('config.json', 'r'))
 
 REPO_URL = 'https://github.com/commaai/comma10k/'
+
 
 # print(config)
 
 @app.route('/js/<path:path>')
 def send_js(path):
-  return send_from_directory('static', path)
+    return send_from_directory('static', path)
+
 
 @app.route('/imgs/<path:path>')
 def send_img(path):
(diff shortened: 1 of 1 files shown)


geohot:
The commapencil tool doesn't work great anyway. Think you can get commacoloring working on this repo, IIRC it was much better.

https://github.com/commaai/commacoloring

nelsonjchen:
Sorry, I don't think I can. To elaborate, I can't preserve the GitHub-oriented collaboration nature of this project where users do pull requests. 

With that, what's left is the UI. 

I'm also not familiar with the comma coloring experience and was never part of it. I'm guessing the distribution of work was better but this pencil tooling's current annotation UI resonates better with me [than the demo I saw for the annotator shoutout](http://kyamagu.github.io/js-segment-annotator/?view=edit&id=1) from a KISS standpoint. It's hard to port something over that you aren't too happy about using.

Some of the other folks in the Discord have gotten some really neat alternative annotation tools going. Especially the ones that are JS devs as a dayjob. Unfortunately, the Git (and GitHub) interaction is still painful.

I *do have* some thoughts on that though so I'm considering of producing a tool for that.

nelsonjchen:
I extracted this tool out to https://github.com/nelsonjchen/tool4comma10k

I tried my hand at fixing pencil, but in the end, I much more enjoy https://github.com/nanamiwang/comma10k/tree/gimp_plugin/gimp_plugin for editing. Whatever the setup to replace the pencil tool is, I hope it includes that plugin!

PR #83, opened 2020-03-05
Update 19-29

malawadd:


VirtuallyChris:
Looking great! Could you just go back and clean up these images? I noticed they have some missing pixels or wrong pixels (particularly some red on the edges of the hood) 
image Thanks!

malawadd:
I have a lot of free time , so yeah i can do that , but i'm not entirely sure what im looking at in the pictues you attached.

VirtuallyChris:
I just zoomed in on where cars meet the road, I noticed some of the pixels aren't right (example above has non-road color under a car before the road color is up)

PR #89, opened 2020-03-05
Mask 0201 png

tracminhvu:
Cleanup edges around objects and road


geohot:
Seeing a conflict, can you rebase on master?

pjlao307:
This is looking pretty good, only thing I would say is mark the lane lines better.  As George stated in discord, those ones that are farther away from the car are important to improve behavior.
image

PR #95, opened 2020-03-05
[500, 501]

madebyollin:
One night frame, one heavily shadowed frame.


geohot:
In 500, I see a leftmost lane line that isn't labelled.

PR #108, opened 2020-03-06
819,823,824,828,829,830,831,832,833

uselessinfodude:
cleaned masks


uselessinfodude:
changes made

pjlao307:
These look good!

PR #118, opened 2020-03-06
818,820,821

uselessinfodude:


pjlao307:
Why does this PR list 3 images but there are 12 files modified?

uselessinfodude:
No idea, I'm guessing it's still sending the files I sent in my last PR even though I created a new branch, edited these 3 files and committed/pushed from that branch.

pjlao307:
You need to rebase master, then create an new branch from master so it has the latest files.
To do that you need to checkout master, then when you create a new branch it will be based off of master, not your other branch that may have different files.

uselessinfodude:
I'm using sourcetree and not exactly sure how to 'rebase' master. The only option I'm seeing is 'rebase current changes onto master' which I believe you said before is not correct.

pjlao307:
Your master is currently 7 commits behind Comma's.  I don't use sourcetree but the discord channel has the commands to do it via command line.

PR #120, opened 2020-03-06
Updated images 161 - 165

scottadair:


scottadair:
164 - It looks correct to me. The driveable goes right to the guard rail 
165 - Didn't know we were masking the reflectors. Will update

pjlao307:
164 - you are correct, for some reason I though that was the road going the opposite direction.

PR #122, opened 2020-03-06
818,820,821,834,835,836,837,838,842

uselessinfodude:


uselessinfodude:
Changes made

pjlao307:
834,835 look good.  818 is still missing the road marking on the left side (past the railing).
image

uselessinfodude:
Made the change although it looks like a curb or edge of road to me

PR #123, opened 2020-03-06
66-70

pjlao307:


tracminhvu:
Hi Pilao307, is mask 066 ok? there is a few pixels (un drivable) near the wiper, but it is right in the middle of the road, and may confuse the model. 
image

pjlao307:
Thanks for catching!
https://github.com/commaai/comma10k/pull/125

PR #127, opened 2020-03-06
masks for [0461-0469]

mobula3967:


VirtuallyChris:
Good stuff! Do you mind if you fix the wide lane line at 468 to split into the individual double line fragments? It won't look as clean but that will probably help out the model a ton if only the yellow are really lane markings

mobula3967:
this message is so cool! makes it much easier to manage the branches

PR #146, opened 2020-03-07
Update to #654 mask

tiki-bug:
re-mask complete


tiki-bug:
made a new commit with the updated mask, stop line removed, should be good to go!

PR #163, opened 2020-03-08
771, 772, 773, 774, 775 - update masks

snevah:
0771, 0772, 0773, 0774, 0775 - update masks in GIMP by snevah


snevah:
fixing 773

PR #179, opened 2020-03-08
0781, 0782, 0783, 0784, 0785 - Update Mask

snevah:
0781, 0782, 0783, 0784, 0785 - Update Mask in GIMP by snevah


pjlao307:
Still getting conflicts.

PR #180, opened 2020-03-08
update masks for 624-640

STKFLT:


pjlao307:
@STKFLT all of your masks (including the new batch 631-640) are the wrong size dimension.  Please fix.

pjlao307:
Also, the fie size is huge compared to the original, they're 800-900% larger in size.

PR #185, opened 2020-03-08
masks [490-494]

mobula3967:


pjlao307:
Better, but for 492, these lines are part of the crosswalk so should also be removed.
image

PR #195, opened 2020-03-09
updated 210, 211, 212, 213

tracminhvu:
masks updated.


tracminhvu:
the branch updated with:  (mask 210)
1. removed unwarranted lane marking
2. fixed truck on the right was missing cabin top

PR #199, opened 2020-03-09
650,651,653

LBJ2000:
650 & 651


pjlao307:
Why is there another PR for the same files?  You should be making changes to the existing PR.  If the existing PR is no longer relevant please close it.

pjlao307:
Also, include the images that are being updated in the title of the PR.

pjlao307:
You may want to re-evaluate the tool or process you're using,  16 commits and it's still failing color check.

pjlao307:
650 still has bad colors, 651 doesn't.

PR #211, opened 2020-03-10
update 215, 216, 217,  218

tracminhvu:
Some spots are tricky! not sure if I got them, especially the cam view in the night.
Send back with instructions if needed.


tracminhvu:
216: fixed truck part masking that was marked as un-drivable
 217: on ramp shoulder should be marked as. un-drivable

PR #225, opened 2020-03-10
Viewer enhancements

gregjhogan:
* left/right arrows for back/forward navigation
* up/down arrows for mask opacity
* m to toggle mask
* q or escape to quit

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,4 @@
+tqdm
+pygame
+numpy
+image
--- a/tools/window.py
+++ b/tools/window.py
@@ -27,7 +27,7 @@ class Window():
   def getkey(self):
     while 1:
       event = pygame.event.wait()
-      if event.type == QUIT:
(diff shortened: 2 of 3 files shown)


pjlao307:
Nice.
A couple other features that would make reviewing faster is a toggle for the mask because somethings it's hard to see how accurate the mask is while it's overlaid on the image and maybe a zoom and pan to look at the details.  This is what I do in photoshop when reviewing PRs.

gregjhogan:
yeah I was thinking the same thing about the mask, so added up/down arrows for mask and `m` for toggle

not exactly sure how to add the zoom/pan, will be a good next enhancement

pjlao307:
Are you able to merge this or would you need me to?

gregjhogan:
I will merge it if CI ever passes! seems stuck

gregjhogan:
github actions seems to be experiencing issues for the last couple hours, but since this PR doesn't touch anything that the tests check I am going to merge.

PR #234, opened 2020-03-11
update 250, 251, 252

tracminhvu:
3 more masks updated, send back if there is any problem. Thanks.


tracminhvu:
fixed bad color #250

tracminhvu:
updated lane lines and reflectors. 
Lane line on far left is barely visible, if  I was driving this car, I would miss it completely.

pjlao307:
Re: far left lane line, yes, it's barely visible but we want the AI to see it even if a human doesn't.  That's the path to superhuman driving agents!

PR #257, opened 2020-03-12
0320-0324

orrrobert:


nelsonjchen:
Uh, the sizes of the masks are changed.

VirtuallyChris:
Yeah :/ fix the sizes and we can reassess!

nelsonjchen:
Ah I also see why there isn't a check run. This modifies the original `img` files, not the masks.

orrrobert:
oops tried to save into imgs instead of masks

PR #283, opened 2020-03-14
[411-416] hassanhibbert (h2code)

hassanhibbert:


hassanhibbert:
I don't think 413 is missing the left line marking. That looks like the edge of the ramp to me. Please review and let me know if you want the line marker there anyway.

I'll take care of the rest of the parked cars.

pjlao307:
413 - you are correct.

hassanhibbert:
Updated the masks. Should be good now.

PR #296, opened 2020-03-14
build job is required for every PR, so run the color check always

nelsonjchen:


nelsonjchen:
This should allow #279 to pass. FYI @erikbernheim .

PR #300, opened 2020-03-14
660-664 remasked

tiki-bug:
remask complete
660 661 662 663 664


tiki-bug:
I went back and forth on this for a while.  I'm pretty sure there's a tree here with one of those wire fence rings around it for protection, which is why it's partially transparent.  Still think the area should be classified movable?

PR #330, opened 2020-03-15
0090 russbb02

russbb02:


russbb02:
This is a resubmission of 0090.  The image in this PR should be the correct size.

pjlao307:
Looks like it's still stuck.
Plus, your size is now correct but there are 2 other issues:
1. The road mask isn't going all the way to the edge of the mask (on both left and right side)
2. Your resizing made the edges anti-aliased so this will not pass the color check

To fix the stuck CI issue I would suggest copying your mask file somewhere, DELETING your fork and re-forking from comma's repo.  Then create a new branch from that, copy your file into that new branch, committing that and submitting a PR.

PR #331, opened 2020-03-15
improve 700 to 702

jsee23:


pjlao307:
In the future, please create an issue and post which images you're working on before working on it so others know not to do those.

PR #355, opened 2020-03-17
convert script for ScaleAI PNGs

pjlao307:
This doesn't work for image 283 because scaleai uses the same color for both vehicles and street signs!


pjlao307:
Closing stale PR

PR #359, opened 2020-03-17
update 341, 342, 343, 344

tracminhvu:
new branch PR to fix issues of changed img instead of mask


tracminhvu:
updated 341, 342, 344

PR #364, opened 2020-03-17
511, 449, 448, 447, 446, 445

energee:


energee:
@pjlao307 Can you give an example of how you would handle cars where you can determine the outline?

pjlao307:
Not sure what you mean by that.  If you're asking about those cars I posted for 511 that sit behind the undriveable poles I would outline and color in the cars first then outline and color in the poles on top of it to cover that part of the car that it obscures.

First Pass:
image

Second Pass:
image

Combine all layers:
image

pjlao307:
If you're asking about 448 in the night scene you just have to guestimate the "blob" of where the cars are based on the lights:
image

PR #387, opened 2020-03-18
945, 946 - Cedricske

cedricske:

PR #390, opened 2020-03-18
947, 948   Cedricske

cedricske:


pjlao307:
945,946 are already submitted in this PR https://github.com/commaai/comma10k/pull/387 which I've already reviewed so I'm not going to review them here.  Please remove them from this PR.

cedricske:
I feel that part of the area you selected above is a ditch or not part of the road. While there is for sure a road on the other side, I think its a divided highway.

Issue #399, opened 2020-03-18
822,839-841,843,844


bigbalthazor:

pjlao307:
https://github.com/commaai/comma10k/pull/401

PR #401, opened 2020-03-18
822,839-841,843,844 - bigbalthazor

bigbalthazor:
Feel free to send any of them back i think I'm going to have to do them in gimp.


bigbalthazor:
#399 issue

Issue #386, opened 2020-03-18
950 - 959 meh-meh


meh-meh:

meh-meh:
PR #408

pjlao307:
https://github.com/commaai/comma10k/pull/408

PR #408, opened 2020-03-19
950 - 959 meh-meh

meh-meh:
Completed. Please give extra scrutiny to 951 and 959.


meh-meh:
Issue #386

meh-meh:
I assumed it was the back side of signage on the other side.

pjlao307:
Ok we'll go with that!

PR #411, opened 2020-03-19
h2code 387-389

hassanhibbert:


hassanhibbert:
ok, fixed line marking. I think i remember i did one in the past that i covered the lane because without covering it; it seemed like it could be another drivable lane but it wasn't. I'll have to go back and find that image.

PR #417, opened 2020-03-19
Fix sample gif mask

techied:
Gif credit @Lambaek


pjlao307:
Can the file be compressed more?  It went from 689KB to 5.59MB

techied:
Done

PR #428, opened 2020-03-19
updated 437, 438,  439

tracminhvu:
Initial PR


VirtuallyChris:
Looking good! On 439 the left side of the road is good, but the right side you mark the black portion of the divider as a lane line. We had a recent discussion to only label the white portion. Fix that and you should be good!
image

tracminhvu:
fixed 439 as requested.

tracminhvu:
fixing bad color, soon!

tracminhvu:
fix.  bad color 439

PR #469, opened 2020-03-21
[670-674] - sting

mobula3967:
Closes https://github.com/commaai/comma10k/issues/370


mobula3967:
I expect lots of remarks on this batch. Many doubts on some labeling and objects.

PR #472, opened 2020-03-21
h2code 434-436

hassanhibbert:
Closes https://github.com/commaai/comma10k/issues/419


pjlao307:
Looks like 434 has bad colors.

hassanhibbert:
yeah, pushing a fix now.

PR #482, opened 2020-03-21
711-719, 945,946 cedricske

cedricske:
Closes https://github.com/commaai/comma10k/issues/407


pjlao307:
Please keep the PRs separate, you're combining files from multiple PRs.

cedricske:
712 - those are uhaul trucks, not the building. 
image

cedricske:
713 - That is not a car. What I labeled was distant cars.

pjlao307:
712 - Ok, the part to the right do not look like uhaul trucks, they look like storage units.
https://rewrite.webselfstorage.com/UserImage/f289dab3-c867-4375-98d2-6c1438473b1c/400/400

713 - That is definitely the back of a truck with reflectors on it.  The size of the mask you have would be equivalent to to the wheels with no chassis.  Look at the size of the car next to it.
image

cedricske:
You're right on the storage unit part, my bad. 

I'll change 713 even though I respectfully disagree as I feel those are reflector posts along the edge and not a truck.

pjlao307:
The perspective is wrong to be reflector posts, they're even across the lane and there are no posts near the OP car.  Wouldn't you see them there as well?

PR #497, opened 2020-03-22
786, 787, 788, 789, 790 - remask

snevah:
786, 787, 788, 789, 790 - remask snevah with gimp
Closes https://github.com/commaai/comma10k/issues/322


snevah:
cars added to 0789.

PR #498, opened 2020-03-22
0005 - Added oncoming lane

### LINKED ISSUE #496: 0005 missing oncoming road mask

meh-meh:
Closes #496


meh-meh:
I think we're splitting single pixels in a blurry image. Unless we want to make masks double the resolution to support that, I think we're both as right as is possible.

pjlao307:
Ok then let's wait till another reviewer chimes in.

meh-meh:
Sorry, accidentally clicked the re-review button.

meh-meh:
I tried to see what you're seeing and updated accordingly, but I'm not sure I've given you exactly what you wanted. This is as close as I can get while still believing my result.

PR #505, opened 2020-03-23
920-924 initial masks

thedude130:
Closes https://github.com/commaai/comma10k/issues/504


pjlao307:
Looks like there's color issues on 922, 924

meh-meh:
921 is definitely not painted lane lines. The contrast with the ditch, combined with motion blur, may make it appear ambiguous, but many rural roads look like this.

PR #512, opened 2020-03-23
750 759

LBJ2000:
Closes https://github.com/commaai/comma10k/issues/438


pjlao307:
I'm assuming this PR is supposed to update 750-759?  If so 753 is missing.  I'll review what's here but let me know if you'll be including 753 as well.

PR #527, opened 2020-03-23
h2code 975-979

hassanhibbert:
Closes https://github.com/commaai/comma10k/issues/517


hassanhibbert:
I'm not totally sure. It seemed too round to be a car. I was thinking it was one of these things https://c8.alamy.com/comp/DF4HAH/propane-gas-tanks-for-lease-or-rent-for-homes-and-business-for-heat-DF4HAH.jpg

flamethrower225:
i wonder if the thing with small wheels is actually a bike and the round thing behind the gate is the propane/gas tank

flamethrower225:
i don't believe its a car though. Wheels are way to small to be a car in my opinion

flamethrower225:
I take my comment back, very convinced it's an old camper

Issue #520, opened 2020-03-23
image 512 issues


spektor56:
Shoulders of road are marked as not drivable and yellow lane line behind the white car on the left is missing.

PR #563, opened 2020-03-25
511-520

pjlao307:


spektor56:
Fixes: #520 

This PR fixes the above issue as well

pjlao307:
512 - There isn't much of the road when it reaches the overpass but I extended it further down where it does widen
image

513 - it's hard to make out where the paint is near that car and you can't make out any line at all in front of it it's just a blur of road blended with the shoulder.  I did extend the lane up to the back of the car though.

514 - That is gravel, but by the new definition of "road" people won't look at you funny if you drove on it or stopped there (just like you stopping next to a curb on the street).

PR #564, opened 2020-03-25
521-530

pjlao307:


pjlao307:
521 - But you wound't drive half on the new and half on the old pavement (both cars in that image are sticking to the new pavement) so that's where we'd want OP to drive as well.  I think as long as the environment looks like this we should mark it that way.  When the pavement is even OP should already know to use the entire lane.

528 - You can (barely) see the wiper and hood show through the reflection that's why I didn't mark the reflection as my car.

VirtuallyChris:
Makes sense man, thanks!!

PR #566, opened 2020-03-25
531-540

pjlao307:


pjlao307:
532 - Not sure what you want changed there?  I did add those tail lights as car and removed the lane line marked on black (tar?)

533 - I don't think that's lane line, that's the white part of the curb that meets the road.  Lines don't go right up to curbs like that.

The rest are updated.

VirtuallyChris:
Yeah I see what you mean about 533, thanks man! I meant that the road looks to ends like it's on a hill like this
image

pjlao307:
If the road ends there those lights in the top right that look like tail lights are flying cars then :D

VirtuallyChris:
Hahaha makes sense xD

PR #568, opened 2020-03-25
014 & 103 - sting

mobula3967:
014: fix poor quality
103: Mark car rack in far left as car, fix left car rear-view as well


mobula3967:
wait 103 still needs improvement of car far right back wheel

mobula3967:
fine. merged already

PR #596, opened 2020-03-29
601-610

pjlao307:


pjlao307:
608 - Are you talking about the white streak under the headlight streak?  I believe that's headlight reflection off the road, not lane lines.

The others are updated.

VirtuallyChris:
Ah yeah you're right, that makes sense. Thanks!

PR #600, opened 2020-03-29
641-650

pjlao307:
649 - Not sure about these lane lines on the left side.  Look like concrete curb/divider maybe?  Can someone else verify?
image


pjlao307:
Updated

PR #619, opened 2020-04-04
275 276 278 284 286 302 303 308 309 311 mask touch-up

snevah:
275 276 278 284 286 issue fix by snevah in gimp


snevah:
well shoot... I think one of the files i uploaded to the main code file, rather than the mask. I've put it in the mask folder, but it is still showing up that I have pull request for it to be added into the main code file. Is someone able to fix this, I appear not to have the skills to do that.

pjlao307:
All 5 masks are in the masks/ folder so you should be fine.

snevah:
302 303 308 309 311 mask touch-up by snevah in gimp

PR #627, opened 2020-04-04
761-770

pjlao307:


pjlao307:
That's not the curb, that's the white part of the curb that meets the road which is technically driveable.

pjlao307:
Resubmit to resolve conflicts: https://github.com/commaai/comma10k/pull/655

PR #635, opened 2020-04-04
841-850

pjlao307:
848 - This was a tough one but I think the solid lines are actually tar lines since they're wiggly so I removed them.  Can someone else confirm?


pjlao307:
Resubmit to resolve conflicts: https://github.com/commaai/comma10k/pull/664

Issue #236, opened 2020-03-11
254,255,256,257,258,259,260


pjlao307:

pjlao307:
https://github.com/commaai/comma10k/pull/237

PR #650, opened 2020-04-05
[209, 217, 232 - 236] Second pass

### LINKED ISSUE #647: [209, 217, 232 - 236] Second pass - meh-meh

meh-meh:
Closes #647


ExoMemphiz:
Not sure if I am correct, but isn't there missing drivable surface on the left side on #236?
Sorry if I'm mistaken. :)

meh-meh:
I was trying to only make changes I was 100% convinced of. I could see some areas that I think are walkways and foliage that might actually be roads, but since the original mask didn't mark them that way, and I wasn't sure, I left them alone. If there's a specific part you feel strongly about, let me know.

meh-meh:
Done

PR #672, opened 2020-04-13
resubmit 921-930

pjlao307:


pjlao307:
Updated
922 - Some of the oncoming car features I can't make out so I left them as is.  That middle "lane" I can't tell one way or another so I left that also.

VirtuallyChris:
Makes sense, thanks for your input!

PR #674, opened 2020-04-13
resubmit 941-950

pjlao307:


pjlao307:
945 - That road edge is correct, that's the white (cement) part of the vertical curb that meets the road.  That median is raised (not flat).  What you outlined is the top part of it that meets the red brick part.

947 - updated

VirtuallyChris:
Oh shoot you're right, that's a tricky one! Thanks man

PR #675, opened 2020-04-13
resubmit 951-960

pjlao307:


pjlao307:
Updated gravel.

Per George, movable means it can move on it's own (i.e. a bike with a person on it is movable but a bike by itself is not, or a trailer by itself)

VirtuallyChris:
Makes sense, thanks!

PR #678, opened 2020-04-13
resubmit 981-990

pjlao307:


pjlao307:
981 - updated
983 - I don't think that's a lane line, if it were it wouldn't be that thick.  See how it is the same color as the concrete up near the truck?

VirtuallyChris:
Yeah I see what you mean

PR #690, opened 2020-04-21
1999

gregjhogan:
Wasn't sure if I should mark the washed out lane lines on the left side as they continue to approach the stoplight.  I know they are there, and you can kind of seem them as you zoom in, but the poor image quality at that distance makes them look much wider than what they could possibly be so I didn't mark them.


gregjhogan:
also decided that thing in the distance in the same lane as the car is a lead vehicle, but not 100% sure (at first I didn't think it was, now I do)

VirtuallyChris:
Good stuff Greg! I think this is a driveway or at least some sort of gravel pull-off that could be marked as driveable.
image
Let me know your thoughts

gregjhogan:
yeah that seems right, I will update it, thanks!

PR #691, opened 2020-04-21
1001-1003

pjlao307:


pjlao307:
Updated

WuerfelDev:
The barrier is marked as road in 1000
Screenshot_20200424-085119__01

PR #697, opened 2020-04-22
h000-h005

pjlao307:


pjlao307:
That "sidewalk" actually looks like a driveway.  The sign on top is for a car repair place and there's no edge to the curb.  I added the lane line though.

VirtuallyChris:
I see what you mean, thanks for updating it :)

PR #702, opened 2020-04-22
1950-1957, 1959-1960

doktorsleepelss:


doktorsleepelss:
1958 came out messed out. Having some trouble with with image labeler saving it correctly. Gonna try to fix in a bit..

doktorsleepelss:
I put the original 1958 back. I'll work on it some other time.

doktorsleepelss:
You sure about 1952? I interpreted that that line as the curb of a sidewalk, not a lane line. I think I see green grass immediately after.  Hard to tell at night.

VirtuallyChris:
Yeah I mean if you can't tell then it could go either way, don't worry about that change then :)

PR #708, opened 2020-04-23
h090 h099 touchup

### LINKED ISSUE #696: [h090 to h099] meh-meh

meh-meh:
Closes #696 

h095 was a bit tough to figure out what the features were on the far left and right. Let me know what you think.


meh-meh:
Dang. Was afraid there might be color checking issues on h095 as I had to hand modify. I'll get a new commit up soon

VirtuallyChris:
Yeah just make sure with editing masks on something like Gimp or Photoshop that you turn off aliasing, it will make your life much easier

meh-meh:
I used paint because I didn't want any fanciness, but I used a brush when I should have just used pencil or fill. Should be good now.

meh-meh:
I made all the changes except for the left line in h090. Looking very closely at it, the mask is properly covering all the visible parts of the lane line. I don't really want to draw lines where there aren't any, but if the direction is that all lines should be solid, I can do that.

PR #709, opened 2020-04-23
h006, h007, h010, h012

erikbernheim:


erikbernheim:
Fixes added

erikbernheim:
Whoops, I should have caught that

PR #718, opened 2020-04-23
Masks h080 to h083

scottadair:
Updated masks


scottadair:
For H082 I think that's a curb not a lane marker

For H083 we are marking humans as movable? I thought it was only vehicles?

pjlao307:
All humans (and animals!) are movable

scottadair:
Hmm, ok. Didn't know that

PR #722, opened 2020-04-23
Masks h084 - h089

scottadair:


scottadair:
h087 - I had a lot of debate here. I had it masked but the more I stared the more it looked like a curb rather than a marking

pjlao307:
It lines up with the lane line beind the car though.

PR #726, opened 2020-04-23
#716 [1035-1040] cecnic1989

cecnic1989:


VirtuallyChris:
Unfortunately there are bad colors in these so I can't review them. This means that you have some sort of aliasing on in your editing program causing the image to have more than the 5 mask colors. I'd recommend going back through in a program like GIMP or Photoshop making sure to use the pencil tool only and disabling aliasing wherever possible. Think pixel art when making these!

cecnic1989:
@VirtuallyChris I was using the pencil tool. I'll try again with GIMP.

PR #723, opened 2020-04-23
h058-h062 - Cedricske

cedricske:

PR #731, opened 2020-04-23
h050-h062 - Cedricske

cedricske:


pjlao307:
Also, this PR has the same masks in PR https://github.com/commaai/comma10k/pull/723.  How are you creating your branches?  From your previous branches?  If so, you shouldn't do that because your subsequent branches will contain changes from other PRs.

PR #741, opened 2020-04-24
[1337] b00fy

b00fy:


pjlao307:
You've git bad colors in the file, make sure the mask only uses the exact colors described in the README.  Make sure antialiasing is turned off in whatever tool you use.

pjlao307:
Closing this as the mask is shifted way off and would need to be redone and there hasn't been an update in 4 days.

PR #749, opened 2020-04-24
[1265] alecg22

alecgiusti:
im not sure why the original isnt showing when I uploaded the new image...? is this good or does it need the original to show up like in the tutorial video?


geohot:
You have a space in the filename before " .png"

@erikbernheim How is the GitHub integration coming? This is why we need it!

alecgiusti:
weird. uploaded again and only showed the edited one again. it worked with 1267 however so i may have just screwed 1265 up somehow. gonna try some more in the morning and get the process working consistently.

pjlao307:
You have 2 masks submitted twice, 1 in the masks/ folder and the other in the folder above it.  All masks should only be in the masks/ folder.

erikbernheim:
@geohot it looks like Github doesn't support implicit auth which would be required for this to be client-side only. https://developer.github.com/apps/building-oauth-apps/authorizing-oauth-apps/ 


I see Cabana is running a small Flask server script to handle this, I assume I will have to do the same. https://github.com/commaai/cabana/blob/master/server.py

pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/893

PR #753, opened 2020-04-24
1601 irfndi

irfndi:


pjlao307:
A few things wrong with this PR

1. 1600 is in the wrong place, all masks should be in the masks/ folder
2. The PR only lists 1601, is this PR supposed to update 7 masks (since there are 7 modified files)?
3 Fix color issues for 1601,1602,1603,1604,1605 ( if they are supposed to be modified)

PR #775, opened 2020-04-24
[1024, 1025, 1028]

exozero7:
Check commit desc for details


pjlao307:
Please fix color issue

PR #798, opened 2020-04-24
[1210-1215] terror

terror:


terror:
> _No description provided._

It seems like image labeler and comma pencil yield color issues for me after editing and saving. Very frustrating...

doktorsleepelss:
Have you used any other program to process the image besides img-labeler? Are you resizing it?

terror:
> Have you used any other program to process the image besides img-labeler? This is a direct save from imgr-labeler? Are you resizing it?

The first commit here was resized after saving from image labeler, which I assumed may be the cause of the issue.

However after editing and saving directly from comma pencil and uploading to here the issue still persists.

pjlao307:
Somehow the road edge has anti aliasing on them..

terror:
> Somehow the road edge has anti aliasing on them.. So the resize probably caused it.


> Somehow the road edge has anti aliasing on them.. So the resize probably caused it.

Tried on windows and it worked, can't be bothered with macOS for this type of stuff

terror:
Submitted 1210-1215
Let me know if I need to change some things.



> Somehow the road edge has anti aliasing on them.. So the resize probably caused it.

pjlao307:
I'd recommend you not work off of your master branch, it will make it difficult to keep track of when you're submitting new work. 

Create separate branches for the batch of masks you want to work on so the PRs are kept separate.

I would suggest copying all of your modified masks somewhere and reverting these changes from your master branch. Create a separate branch for all the files you've done and copy them into that branch, then submit a PR for it.

PR #819, opened 2020-04-24
1565, h009

GarthKiepper:
Reattempt after fixing size from https://github.com/commaai/comma10k/pull/818

First PR. Had some trouble identifying the lane markings. Difficult to distinguish old/new paint. Included reflective markers as part of lane markings.


pjlao307:
Please don't work off of your master branch, it will make it difficult to keep track of what you're submitting in the future. 

Create separate branches for the batch of masks you want to work on so the PRs are kept separate.
I would suggest copying all of your modified masks somewhere and reverting these changes from your master branch (or DELETE your fork and re-fork from comma's repo). Create a separate branch for all the files you've done and copy them into that branch, then submit a PR for it.

GarthKiepper:
Indeed. This was my first PR and I realized my flaw immediately when I went to make a second.

As for the mess of commits, there should be two files here h009 and 1565. Can we pull in just those 2 commits or should I resubmit my PR?

pjlao307:
Fix your master and create a new branch with these 2 files

PR #820, opened 2020-04-25
h009

GarthKiepper:
At the far left there are motion blur lights from a cop car that was previously in frame. I made the editorial decision to include them as a movable object and approximated a blob around it.

Other cop cars were a best effort attempt. I'm assuming the car roof's peak is halfway at the bloom of the lights.

Editorial commentary is welcome!


GarthKiepper:
Good catch! I've added the lane marking.

PR #834, opened 2020-04-25
Resubmit 1030-1034

pjlao307:


pjlao307:
Updated.  1030 was actually for 1031 and 1304 was actually for 1034.

PR #839, opened 2020-04-25
[1043,1044,1049] dizzy

kdizzy07:


meh-meh:
Some duplication occurred due to confusion between spreadsheet and issue creation. This one will will close #743 and close #747

PR #840, opened 2020-04-25
[1076-1079, 1080-1084] nicoleleman

nicoleleman:


nicoleleman:
I used the ImgLabeler tool by ErikB, not sure why there's an issue with the colors?

doktorsleepelss:
Check failed because your edits's resolution is larger than the original. This often happens when you save from img-labeler when you're zoomed in chrome or are scaling within your OS.

nicoleleman:
Got it, so I need to resize them down to 1164px x 874px

doktorsleepelss:
If you're lucky you can resave them them from img-laber by finding them in revert from history under advanced.  

If you choose resize, make sure the program doesn't introduce anti-aliasing otherwise the color check will fail. Image should only have 5 unique colors max.

nicoleleman:
Managed to resave them from img-labeler like you said, will double check the images before submitting next time. Thanks for the help.

nicoleleman:
Re-uploading via my own branch

PR #848, opened 2020-04-25
[1060,1066-1068] dizzy

kdizzy07:


pjlao307:
Need to fix bad colors in masks.  Make sure only colors described in the README are present.

pjlao307:
1060 has bad colors

PR #860, opened 2020-04-25
[1074,1103-1106] dizzy

kdizzy07:


kdizzy07:
I assumed that when you wrote to remove the bike paths that also applies to their lane markings. Let me know if that's not the case.

pjlao307:
Yes, including lane markings since neither path or marking is used to keep the car in a drivable lane

kdizzy07:
Cool, then it's ready for a re-review.

VirtuallyChris:
Looks great! Thanks so much :)

PR #861, opened 2020-04-25
H019, H030, H033

GarthKiepper:


GarthKiepper:
I made the requested changes.

For h030 we appear to be at a 4-way intersection. I think the road ahead goes to the horizon. As you requested, I pulled the road mask it back a bit and cleaned up the borders. Let me know if this looks good or if you think I should pull it back more.

VirtuallyChris:
Looks great! Thanks so much!

PR #862, opened 2020-04-25
1455-1456 piirakka

mardeth:
Only small tweaks needed.


mardeth:
I suppose the tool lets you create new/wrong colors somehow. I'll just close.

PR #864, opened 2020-04-25
h036, h038, h042

GarthKiepper:


GarthKiepper:
Good catch! I've performed the changes you requested and have resubmitted this pull request: https://github.com/commaai/comma10k/pull/872

PR #866, opened 2020-04-25
1151, 1152, 1153

WuerfelDev:


WuerfelDev:
How can I fix that, I used only img labeler in Brave (someone reported it as working)

VirtuallyChris:
Somehow the img labeler has colors that are exactly 1 off from the right ones specked all over the image, I'd report this to the dev of img labeler! (see all the little dots?)
image

What I'd recommend is pulling these into GIMP and using fuzzy select with a composite threshold of like 20 or something, then filling that area with the proper color. This should fix them up.
image

pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/988

PR #874, opened 2020-04-26
Masks - 1305 to 1309

scottadair:


pjlao307:
Doh, looks like the image size changed on all 4 updates.

meh-meh:
@scottadair make sure your browser zoom level is at 100% (assuming you are using img-labeler). You can use 'Ctrl+0' (number, not letter) to reset zoom to 100%.

PR #875, opened 2020-04-26
mask 1719-1735

asterisk-print:


pjlao307:
Also please don't work off of your master branch, it will make it difficult to keep track of what you're submitting.  When you work on other masks in the future they will get included with these masks and it will be hard to manage.

I would suggest copying all of your modified masks somewhere and reverting these changes from your master branch. Create a separate branch for all the files you've done and copy them into that branch, then submit a PR for it.

PR #889, opened 2020-04-26
1874 mask

EgorGorokhov-EG:
Only one just to see if it passes the checks


EgorGorokhov-EG:
> Overall, this looks really good! I would just make a couple small changes:
> 
> The edge of "my car" is a bit weird on the left. On the right, there are road pixels between "my car" and the lane marking, which shouldn't happen, and it seems the left side of the lane marking overlaps "my car". I also think the left lane marking could extend a tiny bit further.
> image

Thanks for your remarks, I've remasked it. Hopefully, it's better now

PR #890, opened 2020-04-26
[1101,1111,1173] dizzy

kdizzy07:


pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/1099

kdizzy07:
sorry for not responding. I was on the fence about 1101. I see how the mark left of the car is just light but on the right you can clearly see the shadows are below my markings so I assumed there had to be something there.

PR #891, opened 2020-04-26
[1570-1575] setrf

setrf:


pjlao307:
Unfortunately your images have bad colors in them, meaning there are more colors than the 5 we use for masks. This happens if you use a program like Gimp or Photoshop without disabling aliasing. Go ahead and check these to make sure there's no aliasing in them, there is often options in Gimp or Photoshop to disable aliasing.

pjlao307:
Closing this so someone else can work on these masks.

PR #902, opened 2020-04-26
[1176-1179] dizzy

kdizzy07:


kdizzy07:
I thought I should account for the curb's height but I see how that's negligible.

PR #903, opened 2020-04-26
1184 masked

NoahWL:
This one was hard since it's a night shot.  There's a lot of motion blur/ghosting with lights which makes it hard to choose a definitive edge to mask.  Below are some parts of the image I'm not completely confident in and think others should weigh in on.

- Tiny red dots in the distance, to the right of the lead car, are likely cars but may be something else
- Additional road to the sides of the main road
- I'm almost certain the red glare on the bottom-right of the windshield is the truck's tail lights reflecting in some fog on the windshield.  I've included what I believe to be the fog in the "My Car" mask.


pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/1148

PR #905, opened 2020-04-26
Masks - 1310 to 1314

scottadair:


scottadair:
1313
- The first one you noted is a person. People are moveable? I've been asked to label them in the past
- The second two I agree with and will update the mask

1314
- Yeah that one was hard. It appeared to be a headlight to be, but I'll take a second look

whichtom:
Ah thats a person? I see it, you're right. The third one is weird but I think you're right here also. Thanks!

PR #907, opened 2020-04-26
1222

HarshaLaxman:


pjlao307:
Unfortunately your images have bad colors in them, meaning there are more colors than the 5 we use for masks. This happens if you use a program like Gimp or Photoshop without disabling aliasing. Go ahead and check these to make sure there's no aliasing in them, there is often options in Gimp or Photoshop to disable aliasing / use the pencil tool at all times will also work.

HarshaLaxman:
Hmm I only used the pencil tool

The commit button didn't work, so I clicked download and moved it to the masks folder

pjlao307:
Pencil tool can be buggy, use https://erikbernheim.github.io/img-labeler/ instead

pjlao307:
Closing this and releasing this so someone can claim it.

PR #911, opened 2020-04-27
1090-1091-1092-1093

userid7:


pjlao307:
See comment in this PR: https://github.com/commaai/comma10k/pull/923
Please designate which PR to review as both of these update the same files.  Close the one you don't want reviewed.

PR #922, opened 2020-04-27
2107

KeeJef:


KeeJef:
This was produced using the pencil tool inside the repo, i suppose it has some bugs since it seems to produce some invalid colors

PR #932, opened 2020-04-27
1269 hsurfer

hsurfer:
mask created with image labeler
img 1269


hsurfer:
the zoom using the wheel mouse resize the image T_T

PR #936, opened 2020-04-28
1098 - 1100 cameron20020

cameron20020:


mutusmurum:
Small fix on your masks:
https://github.com/commaai/comma10k/pull/1045

PR #937, opened 2020-04-28
1270-1280 cameron20002

cameron20020:


mutusmurum:
Fix: https://github.com/commaai/comma10k/pull/1058

PR #938, opened 2020-04-28
[1762] rey2424

rey2424:


pjlao307:
You don't need to close and delete the PR, just commit an updated file to the branch and the PR gets updated.

pjlao307:
We lose the history and comments if you do this.

PR #941, opened 2020-04-28
1868-1870 ellenoize

EgorGorokhov-EG:


EgorGorokhov-EG:
> 1868 - This while section should be undrivable since it's part of the guardrail
> image
> 
> This is a 3 lane highway so the left side the edge of the road goes out to here so that area below should still be road
> image
> 
> Also missing these tail lights as movable
> image
> 
> 1869 - Make the whole car mask movable
> image
> 
> Extend the road out to the grass since that entire shoulder is drivable
> image
> 
> 1870 - This looks great!

Done. But I'm not sure that I fixed 1868 correctly, so could you please check it again.

PR #955, opened 2020-04-29
[1235-1239] mutusmurum

mutusmurum:


mutusmurum:
Not passing color check for some reason.

pjlao307:
This is because there are incorrect colors on each mask:
image

Make sure the tool you use are set to not anti alias and they use the correct color values described in the README.

Here's your color table for 1235:
image

Notice it has all these greyscale colors in the table.  The issue is you have "Custom".  You need to use "Selective" when saving.  This will only uses the colors that are actually in the file and not the entire color palette.

pjlao307:
However for that image I checked your red (lane) and green (movable) are slightly off color.  
It shows f70605 for red and 10f165 for green.

You'll need to fix that.

mutusmurum:
Resubmited: https://github.com/commaai/comma10k/pull/960

pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/960

PR #959, opened 2020-04-29
[1017-1021] nicoleleman

nicoleleman:


mutusmurum:
Great masks. Missed only one lane mark on the 1217. Fixed here: https://github.com/commaai/comma10k/pull/1053

PR #967, opened 2020-04-30
1108 - 1110 cameron20020

cameron20020:


mutusmurum:
Small fixes: https://github.com/commaai/comma10k/pull/1046

PR #981, opened 2020-05-01
[r006 and r008] -- meh-meh

meh-meh:


mutusmurum:
Small fixes: https://github.com/commaai/comma10k/pull/1154

PR #993, opened 2020-05-02
H039,h040,r000 - Cedicske

cedricske:


mutusmurum:
Fix: https://github.com/commaai/comma10k/pull/1198

PR #997, opened 2020-05-03
[1410-1411]  vackyu

vackyu:
My first mask, please review.

The night shot is a bit hard.


pjlao307:
Your image has some bad colors in it. Need to fix them.  Only the 5 colors from the README should be in the mask.

PR #1010, opened 2020-05-04
[1810-1815] whichtom

whichtom:
1814 has some debris on the road it looks like? dunno what to do about it so masked it as undrivable


pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/1259

PR #1017, opened 2020-05-05
1187-1188-1189-1190

userid7:


YosanAI:
I am new myself and might be wrong. but I don't think they like it when you send a pull request for each edited image. instead, try to include all the images with just one Pull request

mutusmurum:
Reviewed:
https://github.com/commaai/comma10k/pull/1047

PR #1026, opened 2020-05-06
[1876-1880] h2code

hassanhibbert:


mutusmurum:
Fix: https://github.com/commaai/comma10k/pull/1279

PR #1048, opened 2020-05-07
[1510-1515] BornAgainGenius

YosanAI:

PR #1069, opened 2020-05-08
1510 to 1514 BornAgainGenius (resubmit)

mutusmurum:
Original PR from BornAgainGenius: https://github.com/commaai/comma10k/pull/1048


spektor56:
1511, top of the SUV with the tire on the back is not labeled
yellow line on left not labeled
1514 has a bit of yellow line not labeled as well
other than that they look good

PR #1086, opened 2020-05-10
r009-r015

pjlao307:


mutusmurum:
Missing some side-view mirrors. But very good masks, didn't need to fix much: https://github.com/commaai/comma10k/pull/1128

PR #1135, opened 2020-05-12
1578-1590 Kogepang

Kogepang:


mutusmurum:
Fix: https://github.com/commaai/comma10k/pull/1188

PR #1145, opened 2020-05-12
1816

spektor56:


mutusmurum:
Fix: https://github.com/commaai/comma10k/pull/1199

PR #1152, opened 2020-05-13
1832

spektor56:


mutusmurum:
Is that a train on the right? I'm not sure.

spektor56:
yes, its a bunch of train cars

PR #1201, opened 2020-05-15
Simplify README

mutusmurum:
More attractive sample image. Remove legacy labels. Simplify help section. Move youtube link to invidio.us.


pjlao307:
Also you shouldn't PR this from your master branch.

mutusmurum:
> Why not just update sample.jpg? The checker failed because there are files in the root folder that it's not expecting.
I didn't know if it was alright or not. I think the first sample was the first image on the dataset... didn't wanted to replace it.
> Also why change the link from youtube to invidio.us?
Just for better privacy. Invidio.us doesn't load javascript and doesn't include trackers. I can change if you wish.
>Also you shouldn't PR this from your master branch.
That's true. Since it was just README, I thought it was ok.

PR #1293, opened 2020-05-20
1913 to 1921 JNLN01

JNLN01:


mutusmurum:
Fix: https://github.com/commaai/comma10k/pull/1301

PR #1296, opened 2020-05-20
1897 1900 cedricske (not 1899)

cedricske:


mutusmurum:
Small fixes: https://github.com/commaai/comma10k/pull/1303

PR #1304, opened 2020-05-21
1460-1470

pjlao307:


spektor56:
in 1469 i think the road goes further to the left and the black edge is just a shadow of a post.  Doesn't make sense for the other lane to be so skinny

mutusmurum:
Fix: https://github.com/commaai/comma10k/pull/1307

PR #1346, opened 2020-05-29
h380-h382, h399

doktorsleepelss:


doktorsleepelss:
added h399

Kind of unsure of the windshield reflection, but I included it because it it's super visible.

pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/1353

PR #1349, opened 2020-05-30
2456 to 2460 logiti

logiti:


doktorsleepelss:
Your masks are the wrong resolution.  If you're using imglaber, make your browser isn't zoomed in when you save, otherwise they'll come out the wrong size. . And if you're using a 4k monitor, don't save with any scaling enabled in your OS.  Also, the imlaber doesn't really work properly with anything other than chrome.

PR #1356, opened 2020-05-31
2007-2009 masks

NHKK:


doktorsleepelss:
If you're using img-labeler, don't be afraid to go past the border.  It automatically cuts off the excess when you save.  It doesn't seem like you're doing this because  there's lots of undrivable pixels at the border around my mycar and on the border where's there's road. Also, don't leave undrivable pixels between mycar and road. they should be perfectly touching.  You do this by stretching the road past mycar, and then simply overlapping mycar over the extra road.   

I also noticed in 2007, a piece of lane is overlapping mycar, when it should be the other way around.  You can fix this by going to layers button, and moving the mycar layer to top.   

This isn't a full review, but just some general advice when working on masks.  Overlapping is very importanting to produce clean maks.

pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/1357

PR #1363, opened 2020-06-01
2450-2454 logiti

logiti:


pjlao307:
This one's got bad colors on all 5 files.

logiti:
what does that mean?

pjlao307:
It means your file has colors that are not the 5 expected as described in the README.

logiti:
I've tried crosschecking and it seems like everything I did is right. Any ideas on what is causing this error?

doktorsleepelss:
Did you use imglaber?  What browser did you use? Did you use any other programs? Did you do any resizing?

PR #1366, opened 2020-06-01
2120 logan

artificial2:


doktorsleepelss:
If you're using img-labeler, don't be afraid to go past the border. It automatically cuts off the excess when you save. It doesn't seem like you're doing this because there's lots of grey undrivable pixels at the border around my mycar and on the left side of the road. Also, don't leave undrivable pixels between mycar and road. They should be perfectly touching. You do this by stretching the road past mycar, and then simply overlapping mycar over the extra road. You do the same with the lane lines. Extend them past mycar, and then simply overlap my car on top. You can move different layers  to the top  and bottom in the layers table.

doktorsleepelss:
Also, don't upload anything to your master branch. Leave mastter alone. You need to create a new branch for every new pull request.  I would recommend deleting your current fork to start from scratch.   Make a new branch, and upload your mask there.

artificial2:
Thank you, I'm new to this. Thanks for the suggestion.

PR #1368, opened 2020-06-02
2120 artificial

artificial2:


doktorsleepelss:
So it looks like you uploaded your mask to master again, and did a pull request from master again too.   Gotta redo it again. 

I see you did successfully make a new branch (artificial2-2120), but your pull request didn't come from there. You also uploaded it in the main folder in addition to the mask folder.  Only upload to the mask folder. 

Also, you did fix the unwanted undrivable parts. But now you have huge chunks of mycar missing.  There should be no gaps between the lane lines and my car. They should be perfectly touching. 

Here's an example of the sort of steps you would need to take to accomplish this. 

part1

PR #1396, opened 2020-06-04
Add files via upload

owewnwnwn:


pjlao307:
A couple things
1. Please name your PR with the mask(s) number
2. There's a bad color in the mask.  In the upper left corner of the image is 1 pixel of cyan color in it that should be removed.

PR #1402, opened 2020-06-04
2285 to 2289

quantumsnow:
2285 to 2289


pjlao307:
The size of these masks changed.   They must be the same as the original masks.  Make sure you're not zoomed in/out when saving masks in img-labeler.

quantumsnow:
Hm, I'm not zoomed. Could it have to do with Windows UI scaling?

pjlao307:
There should be no scaling either.

quantumsnow:
Ah, ok. I'll do that. But that's a shame, I don't know that I'll be able to do more then if I have to switch off scaling every time 😕

pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/1417

PR #1403, opened 2020-06-04
2255 to 2259

strangequack:
2255 to 2259


pjlao307:
The size of these masks changed. They must be the same as the original masks. Make sure you're not zoomed in/out when saving masks in img-labeler.

mutusmurum:
Fix: https://github.com/commaai/comma10k/pull/1421

PR #1419, opened 2020-06-06
2151 to 2155 artificial2

artificial2:


mutusmurum:
Fix: https://github.com/commaai/comma10k/pull/1420

PR #1422, opened 2020-06-07
Erikb h221-h225

erikbernheim:


mutusmurum:
Fix: https://github.com/commaai/comma10k/pull/1424

PR #1430, opened 2020-06-09
2246

vjau:


doktorsleepelss:
If you're using img-labeler, don't be afraid to stretch out the labels past theborders.  It automatically cuts off the excess when you save. It doesn't seem like you're doing this because there's still grey undrivable pixels on the left side of the road and you did not completely fill out mycar at the bottom.  Also, don't leave road pixels between mycar and the lane lines. They should be perfectly touching. You do this by stretching the lane lines past mycar, and then simply overlapping mycar over the excess lane lines. You can move different layers to the top and bottom in the layers table.

vjau:
I updated the mask following review.

pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/1441

PR #1464, opened 2020-06-19
2071 to 2075 ZuluSpl0it

ZuluSpl0it:
Masks for 2071 to 2075


doktorsleepelss:
If you're using img-labeler, don't be afraid draw the labels past the border. It automatically cuts off the excess when you save. It doesn't seem like you're always doing this because in some images there's lots of undrivable pixelsa around the edge of borders . Also, don't leave undrivable pixels between mycar and road. They should be perfectly touching. You do this by stretching the road past mycar, and then simply overlapping mycar over the extra road. Similarly, in some of the images you need to extend lane lines past my car, and overlap mycar over them so there' no road pixesl between.   You can move layers and up and down in the layers tabble to accomplish this.

pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/1473

PR #1671, opened 2020-08-08
Resubmit 2851-2863

pjlao307:
For #1670


doktorsleepelss:
In 2856, I missed the undrivable lightpole on the left.

PR #1704, opened 2020-08-16
3600 Valhalla

JosephVellella:
First 10k PR. Anything I missed?


pjlao307:
Yes, colors are not correct.  Also this has 5 images not just 3600.

pjlao307:
Also all of the masks with cars in them are missing those as movable.
And the My Car edges that meet the road have undriveable strips in them that shouldn't be there.

PR #1713, opened 2020-08-17
Ensure that masks are not cached in browser in pencil tool

Spacefish:
Had the problem where old masked where loaded in the browser, when modifying the .pngs externally.
This fixes this issue, by disabling the browsercache in the development server and appending a timestamp as a query parameter to the image name.


pjlao307:
Closing this since this code isn't being actively developed.

PR #1775, opened 2020-08-26
2756-2760 logiti

logiti:


pjlao307:
There are color issues in all 5 masks that need to be fixed.

spektor56:
First image is missing road and lane line on the right
lanes

PR #1804, opened 2020-09-03
0542-0546 tmilost

tmilost:


pjlao307:
Don't do older ones, only review the ones from the spreadsheet.

tmilost:
My mistake. For the h500 image and above it they say the image does not exist, I thought it was 0500.

PR #1818, opened 2020-09-11
cedricske h580 resubmit 

doktorsleepelss:


doktorsleepelss:
https://github.com/commaai/comma10k/pull/1817

PR #1819, opened 2020-09-12
h542 to h546

WrijeyReaax:


doktorsleepelss:
https://github.com/commaai/comma10k/pull/1821

doktorsleepelss:
Because these images are from the hard set and are not in ideal conditions,  the automatic segnet mask are very unreliable. I would recommend you do these from scratch, and not rely on the default mask.  They will come out way better looking if you do this, and it may not necessarily take you that much more time.

PR #1832, opened 2020-10-01
added masks SA 200-299

Umbrasi:


doktorsleepelss:
Resubmit
https://github.com/commaai/comma10k/pull/1835

PR #1833, opened 2020-10-01
added masks SA 300-399

Umbrasi:


doktorsleepelss:
resubmit
https://github.com/commaai/comma10k/pull/1840

PR #1834, opened 2020-10-01
h620 to h624 LoganxDev

LoganxDev:


doktorsleepelss:
Resubmit https://github.com/commaai/comma10k/pull/1841

PR #1836, opened 2020-10-02
LoganxDev h625 to h629

LoganxDev:


doktorsleepelss:
Resubmit https://github.com/commaai/comma10k/pull/1841

PR #1846, opened 2020-10-08
added masks SA 500-599

Umbrasi:


doktorsleepelss:
https://github.com/commaai/comma10k/pull/1857

PR #1863, opened 2020-10-14
H731-h740-cedricske

cedricske:


spektor56:
image
the cars on left are missing, not sure what is movable in the distance on the right

image
lane line here i can see go all the way to the car infront, can also see the left lane line going a bit further as well

spektor56:
ignore my last comment, i had original masks toggled on lol

spektor56:
Very hard to tell whats going on here but doesn't make sense for there to be a hood bump on the right but not on the left

image

PR #1867, opened 2020-10-15
LoganxDev h661 to h665

LoganxDev:


doktorsleepelss:
https://github.com/commaai/comma10k/pull/1869

PR #1871, opened 2020-10-17
Resubmit h666 h676

FilippoBURRO:
Size fixed


pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/1875

Also, for future reference, most of these masks had issues where you had stray strips of the wrong segments, here are a few examples.

h666 - strips of undrivable along the My Car segment
image

h667 - Same thing on the left, right and bottom
image
image

h670 - along the right side and bottom
image

You need to be careful that you cover all the area with the appropriate mask so you don't leave gaps like thos in the layers.

PR #1872, opened 2020-10-18
LoganxDev h677 to h686

LoganxDev:

PR #1877, opened 2020-10-21
Resubmit h677-h686

pjlao307:
For #1872


doktorsleepelss:
https://github.com/commaai/comma10k/pull/1887

PR #1903, opened 2020-10-31
sa01200 to sa01205 Caddde

Caddde:


doktorsleepelss:
If you're using img-labeler, make sure there's no browser zoom when you save.  Also disable any scaling in your OS if you're using a retina/4k monitor.  Img-labeler can be a little buggy with saving in the correct resolution if you don't.

Caddde:
Thanks, i am going to try to change some thing and retry.

doktorsleepelss:
Also, enter your name in the spreadsheet for the images you're gonna do so no else accidentally works on them at the same time.

Caddde:
Yes i will, i only didn't wrote my name because the check were not successful.

doktorsleepelss:
You need to write your name before you work on them, not after.  If you're gonna work on the same ones, do it right away.

Caddde:
Ok.

PR #1931, opened 2020-11-09
sa01280 to sa01299 Caddde

Caddde:


doktorsleepelss:
https://github.com/commaai/comma10k/pull/1935

PR #1933, opened 2020-11-09
sa1401 to sa1405 ZikoUz

ZikoUz:


pjlao307:
Closing this for https://github.com/commaai/comma10k/pull/1951

Also some feedback:

As doktor said, all of these masks have specks of undrivable along the My Car area that I had to fix.  You need to cover the areas appropriately so they don't leave gaps because those would get filled with undrivable pixels.

Also, don't mark parking spot lines in parking lots as lane lines and you consistently missed reflectors as lane markings.

PR #1975, opened 2020-11-20
1501 1550 fixes and review

cedricske:


doktorsleepelss:
Both this PR has your last one has random polygons all over the place.  Are you by any chance using the Github Integration to commit your fixes? I noticed that that feature only really works when I commit masks that I did completely myself. When I do reviews for existing masks, it doesn't commit correctly.  I just use the regular old save instead, and then commit using the actual github website.

doktorsleepelss:
I'm gonna close your PRs just to declutter a bit. Just submit new ones if you redo them.

PR #1988, opened 2020-11-25
sa01593-sa01596

kunwarsahni01:


doktorsleepelss:
If you're using a 4k monitor/retina display, then disable any scaling in the OS. Also disable any browser zoom.

kunwarsahni01:
@doktorsleepelss should be fixed now.

PR #1991, opened 2020-11-26
Sa01618 to sa01627 Caddde

Caddde:


doktorsleepelss:
resubmit
https://github.com/commaai/comma10k/pull/1998

PR #2152, opened 2020-12-22
6905 6912

jothamsl:


doktorsleepelss:
Check out the video I posted in the discord.  You have lots of undrivable pixels around mycar that shouldn't be there.

jothamsl:
ok

jothamsl:
Is it for 6905 to 6912?

doktorsleepelss:
Yeah, I'm seeing in this for most of them.  And in 6911, undrivable pixels all around the vehicles too.  Gotta overlap everywhere when necessary. And don't leave gaps where there shouldn't be.

jothamsl:
I think 6912 also has the same problem i'll redo everything

doktorsleepelss:
Here's a few more things to watch out for.

6906: The road edge is wrong on both sides. On the right, you caught the edge of the shadow instead of the barrier.  And in the left, the road edge should go past the final lane line where the barrier is. The off ramp should be labeled as road.
6907: Those look like cracks. Shouldn't be labeled as lane lines. Also, no lane lines at the edge of the sidewalk at the gas station.
6908: There shouldn't be any lines here either. And we're labeling driveways as road.
6909:  The road on the right needs to be extended past the lane line to the edge of the barrier.
6910: There's a gap between the barrier and the closest car.  The barrier and the car should be perfectly touching with no road pixels inbetween. You do this by by extending the vehicle polygon into the undrivable.  And the overlap the undrivable on top of the car. 
6912: The road edge is off again here on the right.  The road needs to be extended past the lane line and should be touching the barrier.

jothamsl:
So this is considered as road
image

jothamsl:
And in 6907 there aren't any explicit lane markings should i just make some up.
image

doktorsleepelss:
> So this is considered as road
> image

Yeah, looks like you got the road correct now. That sign is undrivable though, not road.   The onramp behind it should be road though.

doktorsleepelss:
> And in 6907 there aren't any explicit lane markings should i just make some up.
> image

Don't make them up.  Only draw them when you see them.

jothamsl:
Ok, thx, i'll get down to it

jothamsl:
For 6909 the road already extends past the lane mark to the barrier
image

doktorsleepelss:
meant to say left

jothamsl:
yeah i've covered it

doktorsleepelss:
See my changes.
https://github.com/commaai/comma10k/pull/2184

PR #2205, opened 2021-01-03
Update README.md

adhilcodes:


doktorsleepelss:
I think adding the apostrophe makes it possessive.  Maybe capitalisation would make it clearer

`It's 10,000 PNGs of real driving captured from the comma fleet`

adhilcodes:
Yh that will be better I will make PR on that.

PR #2221, opened 2021-01-05
jjmachan - 6955

jjmachan:
This is a sample PR to make sure I've understood all process correctly. I've only filled one image but I'll do it in batch for the next set of PRs.


doktorsleepelss:
See my changes here
https://github.com/commaai/comma10k/pull/2222

PR #2231, opened 2021-01-08
Pledge 1.8.2021 2 images - 8080-8081

cedricske:


spektor56:
size of image 8080 is not correct

cedricske:
The only thing I did differently on 8080 was saved it to a local file before submitting the image. @erikbernheim

doktorsleepelss:
You need to figure something out to get revert from history working. Try using chrome's nightly 
 build and see if they save.  If they do, start using that instead when use img-labeler. https://www.google.com/chrome/canary/

PR #2224, opened 2021-01-06
jjmachan - 6953 & 6954

jjmachan:
did 2 more

PR #2245, opened 2021-01-11
jjmachan 6953 & 6954 resubmit

doktorsleepelss:
https://github.com/commaai/comma10k/pull/2224

See my changes here by pressing space back and forth.
https://spektor56.github.io/comma10kreviewer/?index=0&pr=2245


jjmachan:
so we should include the black markings for lane detection? I was thinking that was not required

doktorsleepelss:
> so we should include the black markings for lane detection? I was thinking that was not required

Yes, we are doing the black parts now.  Originally we didn't, but last month, we decided it was worth including them, and we went back and changed them all.  

Go through the images in the comma10kviewer and for examples of how which black parts to include.
https://spektor56.github.io/comma10kviewer/

PR #2330, opened 2021-01-28
8685 8689

jQuinRivero:


doktorsleepelss:
If you're using img-labeler, you have to make sure you have browser zoom and monitor scaling (if you're using a 4k/retina monitor) disabled. Otherwise it will save in the wrong size.

doktorsleepelss:
https://github.com/commaai/comma10k/pull/2332

PR #2356, opened 2021-02-05
Snow lane fix

cedricske:


doktorsleepelss:
thanks, included these in the resubmit.

PR #2043, opened 2020-12-14
8937-8939 resubmit

hldbrndt:

PR #2444, opened 2021-02-25
baul 8937-8939 resubmit

doktorsleepelss:
fixed bad pixels
https://github.com/commaai/comma10k/pull/2043

changes
https://spektor56.github.io/comma10kreviewer/?index=0&pr=2444


baulml:
nice. thanks for that

PR #2578, opened 2021-04-13
LoftShot 8418-8427

jareducherek:


doktorsleepelss:
Take a look and study my changes here.
https://github.com/commaai/comma10k/pull/2579

PR #2582, opened 2021-04-15
Jaredu 8428-8437

jareducherek:


doktorsleepelss:
Please see/study my changes here
https://github.com/commaai/comma10k/pull/2585

PR #2611, opened 2021-04-29
2607 resubmit

spektor56:
#2607
https://spektor56.github.io/comma10kreviewer/?pr=2611


doktorsleepelss:
In  8482, there's a bit of road on that shouldn't be there on the side of that first tree on the left.

PR #2619, opened 2021-05-02
8381 to 8385 Novus

cedricliu:


doktorsleepelss:
Checks failed because there's too many bad colors, so it can't be merged. It should only be the exact five listed in the readme. You can see that your polygon edges have antialiasing, which is what causing the problem.  Try asking in the discord on how to fix it.   Others in the past have had this problem with photoshop.  

You could also use img-labeler instead, which is probably easier to work with. It's what most people use nowadays.

pjlao307:
If you're using photoshop uncheck the anti-alias option on all selection tools that you use.  Example for the poly tool:
image

cedricliu:
@pjlao307 I see the issue. Is it only work for the lasso tool selection? I tried to create the color block from the pen tool and it does not have the anti-aliasing option for me. Any thoughts?

pjlao307:
I never use pen tool, I only use poly select and fill with foreground or background color.

pjlao307:
Your latest commit should not have changes to the **.gitignore** file.  Please revert that.

PR #2646, opened 2021-05-12
Resubmit PR #2627

pjlao307:


doktorsleepelss:
Why remove the black lane markings from 7687? We're doing those now.

pjlao307:
My bad, feel free to revert it.

PR #2723, opened 2021-06-10
30-e nogira

nogira:
first time; feedback pls


doktorsleepelss:
See my changes here
https://github.com/commaai/comma10k/pull/2724

PR #2791, opened 2021-08-12
00155e-00159e Firebird

Firebird686:


doktorsleepelss:
You need to put the masks in the masks2 folder. You put them in imgs2.

PR #2792, opened 2021-08-12
00155e-00159e Firebird

Firebird686:


pjlao307:
Normally I would just fix these myself but since this was your first time I'll point out some issues I saw so you can take a crack at correcting them and be aware for the future.

155 - You have overlapping undrivable between road and my car.  A good technique to avoid this is to overlap the layers.  For example, extend the road over the hood of the car, then when you add my car layer to it you won't get any gaps that will get automatically filled with undrivable.
image

156 - The car hood needs to be more accurate
image

Also, this is part of the sidewalk so should be undrivable
image

And you have some stray undrivable pixels on the road that should be cleaned up
image

157 - Get more of the far away lane lines and missing some lane lines on the left turn lane
image

158 - Driveways are road
image

159 - Tree trunks and poles are not drivable areas
image

pjlao307:
Also, in case you didn't know, you don't need to close this PR, just update your branch and this PR will get updated.

doktorsleepelss:
> Normally I would just fix these myself but since this was your first time I'll point out some issues I saw so you can take a crack at correcting them and be aware for the future.

Actually, I had already done the fixes already.  So there's no need for Firebird to do them. Just take note of them for next time. 

Firebird, you also left undrivable pixels on the edges of the images.  You could outline outside the border to avoid this. 

[image](https://imgur.com/S3kWmdm.png)

doktorsleepelss:
My changes
https://github.com/commaai/comma10k/pull/2794

PR #2855, opened 2021-10-18
Pyrophoric 172,179 e/f

andrewboldi:


pjlao307:
Normally I would fix this myself but just so you're aware, for 172e/f you're missing the road in the opposite side

image

Of course, the vertical posts will still need to be marked undrivable

image

pjlao307:
Yeah, we want to mark anything we can see because the vision model is used for things like stopped car detection so the sooner it can detect things like that the better the experience.

PR #2875, opened 2021-10-29
Add 192, 193 e/f

PastaPastaPasta:


doktorsleepelss:
193e is the road layer covering the cars.

PastaPastaPasta:
> 193e has the road layer covering everything else.

Good catch! Fixed

PR #2887, opened 2021-11-02
Resubmit PR #2884

pjlao307:


spektor56:
This image has an undrivable border

PR #2890, opened 2021-11-02
Mask 00198F

gudenau:


doktorsleepelss:
See my changes here
https://github.com/commaai/comma10k/pull/2896

PR #2894, opened 2021-11-02
albertorestifo 00133f resubmit

doktorsleepelss:
Overall it was pretty good.  You need to grab the the black parts of  those white lane marking, and you missed that truck on the right side. I'm also more aggressive at grabbing more of the pixels around the lane markings, especially in the the distance. 

You can see my changes in the PR reviewer. Just press space to switch back and forth between your original and my changes.

https://spektor56.github.io/comma10kreviewer/?index=0&pr=2894


albertorestifo:
@doktorsleepelss thank you very much for the feedback! I'll do another single one before starting with batches, to make sure I got it right.

Interesting about the truck: I ended up not highlighting because it's partially covered by a net, so I was wondering what would be needed there.

Also very interesting how you adjusted the driveable area around the "dirt" on the side. Interesting!

PR #2895, opened 2021-11-03
00173f

albertorestifo:
My second attempt at masking, `00173f`


albertorestifo:
I need to make some improvements to one line on the left.

albertorestifo:
Done

Polydynamical (approved):

pjlao307:
I think the part that is in the shadow is undrivable if you compare it to where the shadow is below it (it's in the snow).

albertorestifo:
@pjlao307 had to draw the line there, the is not enough detail.

One on hand, if the part in the shadow is covered in snow, there wouldn't be any space for that right lane that is just starting in the frame, and I find it weird that they cleared the road that way, bulging back inside the road

pjlao307:
I would disagree, you can see where that suv is in the distance in that lane, there is enough room and you can see the edge of the lane where the white (snow) is past the shadow.

image

pjlao307:
For comparison, the proportion of the road when it curves like that is consistent.
image

That part in the shadow is the shoulder.  They wouldn't suddenly have no shoulder going into the turn.

albertorestifo:
@pjlao307 I extended the right side, can you check the changes?

PR #2904, opened 2021-11-11
204e & 204f

odomobo:
First pencil pr. Please provide feedback if any corrections need to be made.


Polydynamical:
Both look very nice. I think you can fill this section in with a movable label because it seems that that area is the back of the bus: 
image

doktorsleepelss:
My corrections.
https://github.com/commaai/comma10k/pull/2905

PR #2908, opened 2021-11-26
Freespirit297 210-214-ef

Freespirit297:
Masks for 210e-214e and 210f-214f.


doktorsleepelss:
Here are my changes
https://github.com/commaai/comma10k/pulls

PR #2921, opened 2022-03-19
Mask for 221e

maxtriano:


maxtriano:
Just realized I forgot to overlap layers in some places, will fix.

PR #2963, opened 2022-05-06
masksd 00024-00029

chrsfr:


doktorsleepelss:
These images have a bunch of wrong colors.  Which browser did you use? Imglabeler doesn't save the images correctly with any browser other than chrome.  It's mentioned in the readme

PR #2977, opened 2022-05-12
imgd images - 9 files

kevinharbin:


kevinharbin:
Are there any issues here? I have more diversity to add once these are approved. In the future, would it be better to upload with file names outside of this convention and let someone with write access pick the appropriate sequence number?

doktorsleepelss:
I normally only merge masks and let comma staff decide whether to merge new images since they're a better judge of what they need in terms of diversity.  But I don't think they're fretting too much about it so I'll just go ahead and merge this for you.

PR #2985, opened 2022-05-17
masksd-49

krkeegan:
This is a tough one, a lot of snow on the windows, which I believe should be "my car".  This is quite the outlier, not sure if that is good or bad.


doktorsleepelss:
Yeah, snow should def be mycar.  We also marked snow on the windshields on the exterior images.

PR #2987, opened 2022-05-19
martinloretzzz masksd 32-33 resubmit

doktorsleepelss:
Try to zoom in a bit closer and be more exact with the outlines.  They were a little bit too jagged and needed to be rounded out.   Missed a few bits and pieces here and there. 

You can see my changes here. Just press space to switch back and forth. Use the arrow keys to get to the next image.
https://spektor56.github.io/comma10kreviewer/?index=0&pr=2987

It's not mandatory, but if you want to save some time, I have some premade mycar masks for the first 48 you can use as a starting point instead of starting from scratch. Just  click on open editor to see them in imglabeler. 

https://spektor56.github.io/comma10kreviewer/?index=0&pr=2947


martinloretzzz:
Thanks for the feedback🚀👍

PR #2996, opened 2022-06-02
KH Variety of dcam

kevinharbin:
empty, driver only, passengers, pet, DM tamper, dusk/night with passengers,


doktorsleepelss:
I would just replace 00068.png with another image of your choosing.  Then I'll feel more comfortable merging.

Maybe something like that would be useful in the future, but the initial segnet is gonna be bootstrapped using only 100 images, and I think this will likely cause it more harm than good.

kevinharbin:
Done and Done. Maybe it'll be useful in the future.

PR #3007, opened 2022-06-15
masksd-62

martinloretzzz:
I've now tried a more complex one, I'm sure there are a few things missing.
I'm not 100% sure about:
- are parked cars moveable?
- cars in the rear window
- where the road ends and what is parking space(is parking space road?)
- the elevated barrier between the roads


doktorsleepelss:
>are parked cars moveable?
Yes

>cars in the rear window
movable

>where the road ends and what is parking space(is parking space road?)
yup driveways and parking space is road

>the elevated barrier between the roads
 undrivable

You can browse through the images  with the image viewer to get a feeling of how how label things.  https://spektor56.github.io/comma10kviewer/


Also,  looks like you used the template for this one.  You can still see grey undrivable pixels where the edge of the window meets the road.  It's annoying, you have to do it this way to avoid the problem.

https://imgur.com/awQNy5A.jpg

PR #3009, opened 2022-07-08
238e TheWonderfulTartiflette

TheWonderfulTartiflette:
This is my first contribution


TheWonderfulTartiflette:
Sorry if I did something incorrectly, that's the first time I do it
By the way, when labelling, if the windshield reflection is too high, near opaque, is that considered "My car" ?

PR #3034, opened 2022-08-07
Manju imgsd 83-87

Saber422:
Thanks for Manju.
Hope my process is Ok to be merged.
Thanks for 10k team.


doktorsleepelss:
These are the wrong resolution and they have black bars.  They should be 1928x1208, but these are 2224x1668.

How are the screenshots being taken?  VLC with snapshot should get the correct size.

Saber422:
OK , I will talk to manju to get correct files, thanks

PR #3035, opened 2022-08-07
Manju imgsd 83-87

Saber422:
My friend Manju used Mac and his VCL player can't catch screen in correct format.
So I make screenshot for him from his hevc files and upload again.
Thanks for Manju.


doktorsleepelss:
looks good

PR #3041, opened 2022-08-11
masksd 91 and 92

AlexandreSato:
this is my first pencil


kevinharbin:
Mask 91:
As mentioned dimensions are off. should be 1928x1208, probably need to start over
Passenger window were the 'movable' meets the 'my car' there needs to be overlap and no 'undrivable' in-between
mask the shade dots at the top of the windshield as 'undrivable'
Some of the driver torso is visible through the steering wheel
recheck all the limits of 'my car' want to be as pixel perfect as we can (e.g. passenger window, rear window)

Mask 92:
Side view mirror on left should mark some road/undrivable between the mirror and car.
Side view mirror on right needs to be marked as 'my car', also the cell phone sticks up near there, mark as 'my car' also
include more of the center scooter front as 'movable'
window sticker as 'my car'
recheck all the limits of 'my car' want to be as pixel perfect as we can (e.g. passenger window, rear window)

PR #3044, opened 2022-08-16
00055 d mask

kevinharbin:
00055 d mask


kevinharbin:
I think this one is ready.

doktorsleepelss:
Colors are bad because it has anti aliasing.

kevinharbin:
I see that :(. Yeah directly from https://spektor56.github.io/img-labeler/

doktorsleepelss:
Yeah, I'm not sure what's wrong. I still have the same problem.

doktorsleepelss:
Is it only this mask?  If you try making a new mask, does it still save with anti aliasing?

kevinharbin:
It's the same build of chrome as before. maybe too many overlapping layers?

doktorsleepelss:
It happens to me with a single layer.

kevinharbin:
Yes, it's happening on another file now.

kevinharbin:
success!

PR #3046, opened 2022-08-17
masks_d_82_to_90

AlexandreSato:


kevinharbin:
Mask 82
I think that's a a parked scooter on the side of the road, mark as movable
Mask 87
Good
Mask 88
Good

kevinharbin:
82 good update
89 looks good

kevinharbin:
Mask 90
small overlap issue with hand on top of steering wheel and movable. has a few pixels of undrivable.
Looking in reviewer with brightness turned up, i think I would trace the top of the child's head differently. I would mark more as movable.
I think the space at the drivers window in front of the parked car is road rather than 'mycar'
I'm not sure about the ground under the scooters and other vehicle. Can more of that be considered 'road'? @doktorsleepelss ?

PR #3048, opened 2022-08-18
00044 dcam mask - Edge

kevinharbin:
Export from Edge, check colors


doktorsleepelss:
In the wrist area of the right arm, you extended interior movable a little bit too much.

kevinharbin:
dang it. I fixed that once. I'll take care of it tomorrow.

PR #3054, opened 2022-08-20
add masks dcam [78-81, 93-95]

AlexandreSato:


kevinharbin:
mask 94 & 95, check the clipping where movable meets my car. I usually overlap the behind layer and use the top layer to define the crossing.
example from 94
image

doktorsleepelss:
You have lots of grey pixels on the edge of the windows. Looks like you used my templates.   To use these, you have extend the road, movables, and lane lines into mycar.  Then you you do a final outline of the window with a mycar layer. 

Something like this.
https://imgur.com/awQNy5A.jpg
 Notice how  drew the road past the window.  Then there's a mycar outline to cover the extra up.  This is how your avoid the grey pixels.

PR #3056, opened 2022-08-22
add masksd 73-77

AlexandreSato:


doktorsleepelss:
In 77, make the driveways and where the cars are parked part of road.

PR #3059, opened 2022-08-24
add masksd [67, 69-72]

AlexandreSato:


doktorsleepelss:
On the passenger side of of 72, I think that's the sidewalk with grass, not part of the road.

PR #3062, opened 2022-08-26
Fixed my car, where part of a chair was in the way

calledit:
Fix obstructing chair in the back window to the left of the image.


doktorsleepelss:
Thanks for giving it a second look.  Now that I look it,  this one is actually missing a number of things, like the windshield.  I was too hasty with this one. I need to make some more changes myself.

PR #3070, opened 2022-08-29
masksd 98 timjeparker

timjeparker:


timjeparker:
Thanks for merging it and for the feedback. It definitely takes a lot of effort, at least on first one, to get things layering cleanly - looking back at some of other image submissions I realised how much effort must go into these. 

Regarding the cars, I am also not 100% sure but if any suggestions for changes I can always make them if doable. I found that even with zooming in/out I just couldn't make out what are vehicles/not vehicles in some of the background .. same with drivable space ... so did a best estimate. I'd like it to be 100% if possible but not sure if there are other ways to confirm what is in the image in some way .. like support detail footage, although I suspect would be based on what is available in the image posted.

Thanks again and great that image count is now up to 100 and training expected to start soon.

PR #3080, opened 2022-09-10
My car templates to masksd 101-110

AlexandreSato:
You can use these as a starting point Instead of starting from scratch to save some time. Wil still require you to redo a lot of it though. Just click on "open editor" in the corner to open them in img-labeler.

https://spektor56.github.io/comma10kreviewer/?index=0&pr=3080

To use these, you have extend the road, movables, and lane lines into mycar. Then you you do a final outline of the window with a mycar layer.Something like this.

https://imgur.com/awQNy5A.jpg

Notice how I drew the road past the window. Then there's a mycar outline to cover the extra slack.


AlexandreSato:
You can use these as a starting point Instead of starting from scratch to save some time. Wil still require you to redo a lot of it though. Just click on "open editor" in the corner to open them in img-labeler.

https://spektor56.github.io/comma10kreviewer/?index=0&pr=3080

To use these, you have extend the road, movables, and lane lines into mycar. Then you you do a final outline of the window with a mycar layer.Something like this.

https://imgur.com/awQNy5A.jpg

Notice how I drew the road past the window. Then there's a mycar outline to cover the extra slack. I

PR #3082, opened 2022-09-13
maskds 101-103

doktorsleepelss:


kevinharbin:
Mask 102, the layers on top of the driver's head got out of order. Out the driver window, I think that's a rack on top of that vehicle,

AlexandreSato:
image
102

doktorsleepelss:
thanks, fixed.

PR #3084, opened 2022-09-14
Add masksd 105-107

AlexandreSato:


AlexandreSato:
> right behind his neck, is that 'undrivable' on purpose or a not quite overlapped layer? I would have gone with 'my car'. Double check, whatever you think is right. image

Got catch kthx

PR #3088, opened 2022-09-19
add masksd 108-110

AlexandreSato:


kevinharbin:
Driver's hand through the wheel should be marked

AlexandreSato:
Is it not the steering wheel?  from the other pictures it appears to be.



Em 19 de set. de 2022, à(s) 16:06, kevinharbin ***@***.***> escreveu:

﻿

Driver's hand through the wheel should be marked

kevinharbin:
I'm looking at this. Looks hand colored.
image
image

kevinharbin:
109, is this a pole in front of the truck?
image
I think this is a raised curb so not actually drivable. Take a second look and you make the call.
image
no comments on 110

AlexandreSato:
> 109, is this a pole in front of the truck?
> I think this is a raised curb so not actually drivable. Take a second look and you make the call. 
 no comments on 110


I think you're right, but I'm not sure. Anyway, I've already made the changes, if anyone else can confirm we'd appreciate it.

PR #3090, opened 2022-09-22
00096 dcam modify mask

kevinharbin:


kevinharbin:
I extended the 'movable' mask here
image

doktorsleepelss:
I think the black and yellow thing is a pole, not part of the bikes.

PR #3093, opened 2022-09-25
Add masksd 113-114

AlexandreSato:


doktorsleepelss:
On 113, there's an concrete barrier right next to the yellow lane line on the driver side. The giveaway is those black dots, which are openings you often seen on these barriers. 

https://imgur.com/OXuNymB.jpg

AlexandreSato:
> On 113, there's an concrete barrier right next to the yellow lane line on the driver side. The giveaway is those black dots, which are openings you often seen on these barriers.
> 
> image

Good catch!

PR #3099, opened 2022-10-03
nworby 131-135

nworb-cire:


doktorsleepelss:
I like that you have multiple people in the car.  Don't get many of these. But can you also include a few with cars driving next to you?  I think that might be important since the segnet  gonna being use for lane changes.

PR #3144, opened 2022-11-18
146-200 mask templates

doktorsleepelss:
You can use these as a starting point instead of starting from scratch to save some time.  Will still require you to redo a lot of it though. 

https://harbn.com/comma10kreviewer.html
Select 3144 on the side, and then you use the  arrows keys untill you find the image you want to work on.  Just click on "open editor" in the corner to  open the mask in img-labeler. 

To use these, you have extend the road, movables, and lane lines into mycar. Then you you do a final outline of the window with a mycar layer. Something like this.

https://imgur.com/awQNy5A.jpg

Notice how I drew the road past the window. Then there's a mycar outline to cover the extra slack.


doktorsleepelss:
hmm.. 151-155 aren't all aligned. I'll make new mass for those soon.

PR #3254, opened 2023-01-22
267f

pbkompasz:
It's my first mask, so please let me know if it is okay.


doktorsleepelss:
My changes here
https://github.com/commaai/comma10k/pull/3256

PR #3267, opened 2023-01-29
267e

pbkompasz:


doktorsleepelss:
There's a ranodm blue line in your mask. I think it's casued by having an open polygon and saving it.

PR #3317, opened 2023-04-14
chs172 416-420

Saber422:
thanks for chs172


doktorsleepelss:
Looks like you 416-420 for this PR too.
https://github.com/commaai/comma10k/pull/3322

Can you rename these?

doktorsleepelss:
Yup, correct.

PR #3319, opened 2023-04-17
00351d lachezartodorov

lachezartodorov:


doktorsleepelss:
smooth

If you zoom in, you'll see how edges are smoothed out, and it introduces new colors outside of the 6.   Unfortunately, there's no easy way to fix this, and you would have to redo it. I'll update the readme to tell people only to use Edge from now since it seems to be broken with chrome right now.

Looks otherwise mostly good.  Even if the cable is being moved, I'd probably make that part of mycar, not inside movable.  Also, you need to catch more of his head. If you increase the brightness (middle slider) and turn off the mask (press space) here, you could more easily see the head's outline.

https://harbn.com/comma10kreviewer.html?index=0&pr=3319

lachezartodorov:
I was wondering about the cable. Thanks for the hair suggestion.

lachezartodorov:
@doktorsleepelss yup on Edge things looks straight.

PR #3356, opened 2023-08-03
Maximize Space and bump up GitHub Actions workflow versions

nelsonjchen:
GitHub Actions was failing due to checkout causing an out of space error. 

Added https://github.com/marketplace/actions/maximize-build-disk-space

Using this GitHub Action I've added, the checks will work again. More levers to trade time for more space are available in that action if desired.

Threw in a bump to some versions of other actions to Node 18 and so on.

Closes #3355 which was a test of the GitHub Actions system to see it broken.

### DIFF
--- a/.github/workflows/check_colors_sizes.yml
+++ b/.github/workflows/check_colors_sizes.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
     - name: Maximize build space
-      uses: easimon/maximize-build-space@master      
+      uses: easimon/maximize-build-space@67afc2d7662c9d296547435e7605cc57a1df46f3      
     - uses: actions/checkout@v3
     - name: Set up Python 3.8
       uses: actions/setup-python@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/check_colors_sizes.yml
```diff
@@ -11,9 +11,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - name: Maximize build space
+      uses: easimon/maximize-build-space@master
```
incognitojam, line 15: Can you pin this to a commit? We like to do that for third party actions

nelsonjchen: Ah crap I didn't realize that was `master`. Sure thing.

nelsonjchen: Done.

nelsonjchen:
image

@doktorsleepelss , you may be able to push to the open PRs of interest by others and rebase them yourself. This checkbox is checked by default for me, so I assume it's also the same for other PRs. A bit surgical, but possible.

PR #3378, opened 2023-10-07
fson 521-525

Saber422:
thanks for fson


doktorsleepelss:
Can you make a new PR with new file names? 521-525 was used for this PR https://github.com/commaai/comma10k/pull/3375.

PR #3395, opened 2023-11-13
Masksd 551~555

AlexandreSato:
- [x] masksd 551
- [x] masksd 552
- [x] masksd 553
- [x] masksd 554
- [x] masksd 555


YassineYousfi:
Looks great, thanks!
Is 553 missing a lane line by the barrier?

AlexandreSato:
> Looks great, thanks! Is 553 missing a lane line by the barrier?

Good catch, fixed! Thanks for review

